PR #1, opened 2018-10-06
Bug fixes

jgiacomini:
- Fix a bug when user go back without taking a picture (AndroidCustomCamera solution)


jgiacomini:
In this PR il try to fixe all small bugs I find in Uno.Samples. :)

PR #26, opened 2020-09-14
feat(ToyCar): Add car animation depending on the hinge angle value and add Uno Platform theme

agneszitte-nventive:
Add car animation depending on the hinge angle value.
Moving the cars in the middle of the device if the angle is equal to 180 degrees.
Also moving them to the right side of the device if the angle is equal to 360 degrees meaning that both screens are facing aware of each other and the device is fully opened. In addition, At the same time, starting the car animation and the scene animation in the background.

Testing the hinge angle with the Surface Duo emulator: https://devblogs.microsoft.com/surface-duo/surface-duo-testing-tips-tricks/#hinge-angle

ToyCar-Android-SurfaceDuo-HingeAngle


UI/ToyCar/ToyCar.Droid/ToyCar.Droid.csproj
```diff
@@ -47,9 +51,12 @@
     <AndroidLinkMode>SdkOnly</AndroidLinkMode>
     <AndroidCreatePackagePerAbi>true</AndroidCreatePackagePerAbi>
     <EmbedAssembliesIntoApk>true</EmbedAssembliesIntoApk>
-    <AotAssemblies>true</AotAssemblies>
-    <EnableLLVM>true</EnableLLVM>
+    <AotAssemblies>false</AotAssemblies>
```
MatFillion, line 54: was it intentional to remove AOT?

agneszitte-nventive: No I think it was changed by going to the properties and I didn't notice the changes at first.
I will update the .csproj thanks ;) !

jeromelaban: It changes back by itself in some editions of VS.

jeromelaban (approved):

UI/ToyCar/ToyCar.Droid/ToyCar.Droid.csproj
agneszitte-nventive: Yes that's what I noticed

agneszitte-nventive: Update done thanks !

PR #39, opened 2021-01-14
feat: Update SkiaSharp Wasm sample to .NET 5

jeromelaban:


nor0x:
@jeromelaban is it possible at this point to use `SkiaSharp.Views.Uno` for the sample? Or does some issues in the offical `SkiaSharp` repo block this? (i.e https://github.com/mono/SkiaSharp/pull/1550)

jeromelaban:
@nor0x for .NET 5 it's not enough, there are additional modifications that need to be made (multi-emscripten compilation) that we'll add to the official repository to enable .NET 5 support.

PR #42, opened 2021-02-23
docs: Adding Time Entry samples

darenm:
This commit adds 3 solutions - 1 Silverlight, 1 Uno  and one WebApi solutions


davidjohnoliver:
The CI doesn't like the Silverlight bits. ;)

Can you move them out of the `UI` directory so they're ignored - you can create a new directory called `reference` and put them under that one.

davidjohnoliver (approved):

PR #61, opened 2021-04-11
Fix CORS error when launching UnoChat.Service locally.

Rytisgit:
Change to use UseCors step earlier in UnoChat.Service configuration so it works when launched locally.
Based on: https://stackoverflow.com/a/60469784

Was not working for me when I launched the service locally. Currently this uses an azure website as a chat hub which hides this issue. https://github.com/unoplatform/Uno.Samples/blob/5564c8f56dd41681e6b9213aa197de28a07bcee3/UI/ChatSignalR/UnoChat.Client/UnoChat.Client.Shared/ViewModel.cs#L71

When reverting to use localhost the problem appears.

### DIFF
--- a/UI/ChatSignalR/UnoChat.Service/Startup.cs
+++ b/UI/ChatSignalR/UnoChat.Service/Startup.cs
@@ -44,12 +44,12 @@ namespace UnoChat.Service
                 // The default HSTS value is 30 days. You may want to change this for production scenarios, see https://aka.ms/aspnetcore-hsts.
                 app.UseHsts();
             }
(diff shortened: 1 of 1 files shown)


jeromelaban:
@carldebilly

carldebilly:
It's strange because on my side it's working both before and after your changes.  Anyway, it's an acceptable fix.

Rytisgit:
@carldebilly thanks, just checking - when you tested did you change the hub urls and use the wasm head to test? I guess I could have been more clear that I was mainly using wasm. The console app worked both ways for me as well.

PR #66, opened 2021-06-15
Uno + Ethereum Blockchain

Onotseike:
1. EthereumSmartContracts- Codebase for the Smart Contracts written in Solidity & JavaScript

2. UnoPharmaSupply - Codebase for Uno cross platform Ethereum blockchain class library written in C# targets Android, iOS, MacOS, Web Assembly & UWP

3. DemoApp - Codebase for Uno cross-platform application that consumes the above-mentioned class libraries. written in C# & XAML


UI/UnoEthereumBlockChain/DemoApp/DemoApp/DemoApp.Droid/DemoApp.Droid.csproj
```diff
@@ -55,7 +55,7 @@
     <Reference Include="Mono.Android.Export" />
     <Reference Include="mscorlib" />
     <Reference Include="PharmaSupply">
-      <HintPath>..\..\..\UnoPharmaSupply\UnoPharmaSupply\bin\Debug\monoandroid10.0\PharmaSupply.dll</HintPath>
+      <HintPath>..\Pre-BuiltDLLS\monoandroid11.0\PharmaSupply.dll</HintPath>
```
carldebilly, line 58: Why aren't you using the project dependency instead?

Onotseike: > Why aren't you using the project dependency instead?

Hi @carldebilly because they need to be built before they can be linked and the git ignore excludes bin and obj folders.

davidjohnoliver: @Onotseike the normal pattern here would be to use the `ProjectReference` tag to indicate the dependency on the local PharmaSupply project. The compiler understands from this that it needs to build PharmaSupply first and include it, no need to include pre-compiled dlls (which should be avoided if possible).

Eg like this:
https://github.com/unoplatform/Uno.Samples/blob/5564c8f56dd41681e6b9213aa197de28a07bcee3/UI/UnoContoso/UnoContoso.Service/UnoContoso.Service.csproj#L12-L15

Could you use `ProjectReference` here, or did you try it already and run into an obstacle?

UI/UnoEthereumBlockChain/DemoApp/DemoApp/DemoApp.Shared/TabbedPages/ManufacturerPage.xaml
```diff
@@ -0,0 +1,393 @@
...
+                               Grid.Row="0"
+                               Grid.Column="1"
+                               Text="{Binding DepolyReceipt.TransactionHash, Mode=TwoWay}"/>
+                            <TextBlock x:Name="contractHashOne"
+                               Grid.Row="1"
+                               Grid.Column="1"
+                               Text="{Binding DepolyReceipt.ContractAddress, Mode=TwoWay}"/>
+                        </Grid>
+
+                        <!--<Grid Padding="20,0"
```
davidjohnoliver, line 158: Can the commented code here be removed?

UI/UnoEthereumBlockChain/DemoApp/DemoApp/DemoApp.Droid/DemoApp.Droid.csproj
davidjohnoliver: Ok, I see now that you have the PharmaSupply project in a separate solution (.sln). You can have as many projects (.csproj files) in a single solution as you want, and in this case to use `ProjectReference` you'll need to put the PharmaSupply class library project in the same solution as the application code and projects. Can you try that?

Onotseike: > Ok, I see now that you have the PharmaSupply project in a separate solution (.sln). You can have as many projects (.csproj files) in a single solution as you want, and in this case to use `ProjectReference` you'll need to put the PharmaSupply class library project in the same solution as the application code and projects. Can you try that?

Hi @davidjohnoliver I have replaced the Prebuilt dlls with the ProjectReference for each csproj. Thanks for the tip.

carldebilly (approved):

PR #108, opened 2022-04-07
chore: SQLiteSample Uno 4.1.9

iury-kc:
SQLiteSample updated to  Uno 4.1.9


jeromelaban:
Have you validated that it does what it's supposed to do? Newer versions of SQLite may not be working properly.

iury-kc:
> Have you validated that it does what it's supposed to do? Newer versions of SQLite may not be working properly.

Yes. It saves the data propperly. After reopen the app, it's still there. Squashing now the new splash screen logo

jeromelaban (approved):

PR #128, opened 2022-04-28
fix: adjusting ports to avoid conflict

### LINKED ISSUE #127: Wrong port in Contoso sample
Hello

UnoContoso.Service is configured to run on port 2706, but App.xaml.cs (in method UseRest) uses port 5000.
It should be changed to:

#if __ANDROID__
            var repository = new RestContosoRepository("http://10.0.2.2:2706/api/");
#else
            var repository = new RestContosoRepository("http://localhost:2706/api/");
#endif

iury-kc:
closes #127 

Ports from WASM and Services projects were the same if using Kestrel.


jeromelaban (approved):
Thanks!

PR #131, opened 2022-06-15
feat(XamlBasics): Add ThreeLevelListDetailSampleApp

lukeblevins:
Adds relevant sample for Xaml Basics writeup [write up](https://platform.uno/blog/xaml-fundamentals-for-web-and-mobile-three-level-list-detail-pattern/)


lukeblevins:
This is the sample for https://platform.uno/blog/xaml-fundamentals-for-web-and-mobile-three-level-list-detail-pattern/

nickrandolph:
@lukeblevins a couple of comments:
- Is there a reason we've only included the Win/Wasm targets?
- Let's see if we can rework the logic to avoid as much code behind and just use data binding where possible
- Let's fix the CI to make sure the sample is building

nickrandolph:
@lukeblevins what's left to be done on this PR?

PR #153, opened 2022-09-20
chore: Upgrade - WebRTC.Wasm Sample to run with latest uno and latest VS 2022

aesteves900:


jeromelaban:
Make sure to use the name of the sample you're upgrading in the PR title. Also, it looks like we missed previously missed this sample for the upgrade (or that we missed it when it was first integrated). Good that we're updating this one now :)

aesteves900:
@jeromelaban , Okay I will do one PR for each integration. Ad back to the first to do "exclusion" to the legacy folder. Despite that, I'm stuck with this error: I will time-box, update and test for two hours, document the issues, and move away to the next when consistently get no success. Sounds good the plan?

image

jeromelaban:
Where is this error coming from? Is this for this sample specificall? Can you show the complete error ?

UI/WebRTC/WebRTC/WebRTC.Shared/App.xaml.cs
```diff
@@ -171,8 +171,6 @@ private static void InitializeLogging()
 				// Debug JS interop
 				// builder.AddFilter("Uno.Foundation.WebAssemblyRuntime", LogLevel.Debug );
 			});
-
-			global::Uno.Extensions.LogExtensionPoint.AmbientLoggerFactory = factory;
```
jeromelaban, line 175: Why did you need to remove it? This is in the original template, and we may need to adjust that piece.

aesteves900: I back up the code and did the required adjustments.

jeromelaban (approved):

PR #156, opened 2022-09-27
chore: Upgrade - WCTDataTreeTabSample to run with latest uno and late…

aesteves900:
…st VS 2022


jeromelaban:
Make sure to update all the projects of the solution

jeromelaban (approved):

PR #160, opened 2022-09-28
feat: AdvancedXBind Sample

peterfoot:
Sample for Advanced x:Bind Blog post.


MartinZikmund (approved):

peterfoot:
I have seen this error but thought it was my development environment. It builds for Windows but I get this error on other platforms. The method exists and is public so I don't understand cause of the error.

kazo0:
Interesting, maybe a bug here: the build seems to pass if you change the binding to `Click="{x:Bind DisplayPlanetSync}"` instead of `Click="{x:Bind DisplayPlanetSync()}"` (remove parentheses from the method name)

kazo0:
Although there does seem to be some issue with styling on Uno platforms:

Skia.GTK:
image


And it seems to be stuck in some sort of infinite loop for WASM, I'm just stuck on a blank screen and nothing loads in the Dev Tools window

peterfoot:
I've updated to remove the parenthesis and it now works on Windows and Uno Android (my dev environment is broken for Blazor at the moment). I've noticed something and I think this should probably be raised as a bug in Uno:-
For a button click handler Windows will accept either with or without the parentheses but Uno errors if the parentheses are present.

```<Button Click="{x:Bind DisplayPlanetSync}"/>```

For a text binding both Windows and Uno accept empty parentheses for a method with no arguments. 
```<TextBlock Text="{x:Bind GetResultOfAsyncMethod()}"/>```
If the parentheses are omitted the text field contains the method signature and not the result of the method on Uno. 
```<TextBlock Text="{x:Bind GetResultOfAsyncMethod}"/>```
On Windows omitting the parentheses causes a build error:-
```WMC1121	Invalid binding assignment : Binding to methods is not supported```

**Comment by kazo0** [comment]
If you can get a simple sample project that reproduces the bug and open a new issue on the Uno repo that'd be great!

**Comment by MartinZikmund** [comment]
Hi @peterfoot, the build is failing due to:

```
error : Building for the iOS Simulator requires the use of the static registrar. Make sure that `MtouchExtraArgs` contains `--registrar:static`. See https://github.com/unoplatform/uno/issues/9430 for more details.
```

This change is required for the csproj: https://github.com/unoplatform/uno/blob/release/stable/4.5/doc/articles/get-started-wizard.md/#dont-know-how-to-marshal-a-return-value-of-type-systemintptr

peterfoot:
@MartinZikmund I've added this to the project file and tested on the simulator. Should this be added to the new project templates? I see that the macos build has this arg added out of the box.

MartinZikmund:
Actually it should be part of new project templates already (Jérôme added it around last release timeline)

PR #163, opened 2022-09-29
chore: Upgrade - Pharma Supply Sample to run with latest uno and late…

aesteves900:
…st VS 2022


jeromelaban:
It looks like the macOS head is missing? Is it not part of the solution?

jeromelaban (approved):

PR #167, opened 2022-10-03
chore: Upgrade - CatApi Sample to run with latest uno and latest VS 2022

aesteves900:


jeromelaban:
There are two projects here as well

jeromelaban:
Looks like there are still two samples ?

jeromelaban (approved):

PR #168, opened 2022-10-03
chore: Upgrade - StatusBarColor Sample to run with latest uno and latest VS 2022

aesteves900:


jeromelaban:
Looks like there are two samples here :)

PR #173, opened 2022-10-04
chore: Upgrade - NativeFrameNv Sample to run with latest uno and late…

aesteves900:
…st VS 2022


jeromelaban:
Not all projects of the solution were updated (Tizen)

PR #178, opened 2022-10-04
chore: Upgrade - EmbebededResource Sample to run with latest uno and …

aesteves900:
…latest VS 2022


jeromelaban:
In order to make the reviews easier, can you open PRs as draft, then put those as ready when the build passes ?

jeromelaban (approved):

PR #194, opened 2022-10-21
Adding SimpleCalculator solution to Samples

eriklimakc:
First commit of SimpleCalculator using Uno.


francoistanguay (approved):

UI/SimpleCalculator/SimpleCalculator.UI/Views/CalculatorMarkupView.cs
```diff
@@ -0,0 +1,253 @@
...
+    {
+        public CalculatorMarkupView()
+        {
+            InitMarkupView();
+        }
+
+        private void InitMarkupView()
+        {
+            this
+                .Background(Microsoft.UI.Xaml.Resource.Theme<Brush>("BackgroundBrush"))
```
dansiegel, line 24: This will need to get updated. Since this needs to be fully qualified on Android this means we need to rename the API.

dansiegel (changes requested):

francoistanguay (approved):

PR #196, opened 2022-10-25
chore: TabView Sample - Update to net6

aesteves900:


jeromelaban:
For this one, use the same technique as the we discussed. Remove all the existing heads, except the shared project, then replace with the new winui heads.

jeromelaban:
This sample does not demonstrate the WCT TabView anymore, but shows the normal TabView. Let's remove it completely, it's not serving much use.

aesteves900:
@jeromelaban , the sample was removed.

nickrandolph (approved):

PR #207, opened 2022-11-01
chore: fix update sample calculator

aesteves900:


nickrandolph:
@aesteves900 I don't think this is required any more?

PR #215, opened 2022-11-06
chore: UnoWCTDataGridSample - Update to net6 #196

aesteves900:


nickrandolph:
@aesteves900 I've pushed this back to draft as the sample doesn't appear to work, at least for the Windows target it didn't show anything other than the column headers:
- Please update to latest Windows App Sdk version
- Can you add in the WPF/Gtk/Linux targets, unless they're not supported?
- Can you fix the sample so it shows data

nickrandolph (approved):

PR #219, opened 2022-11-09
chore: SQLiteOneDriveInvoice Sample - Update to net6

aesteves900:


UI/UnoSQLiteOneDriveInvoiceSample/Demo/Demo/Demo.Droid/MsalActivity.cs
```diff
@@ -1,26 +0,0 @@
...
-using System.Text;
-
-namespace Demo.Droid
-{
-    [Activity(Exported = true)]
-    [IntentFilter(new[] { Intent.ActionView },
-        Categories = new[] { Intent.CategoryBrowsable, Intent.CategoryDefault },
-        DataHost = "auth",
-        DataScheme = "msal4f554894-133f-44c9-92fe-bdcb164ddaa0")]
-    public class MsalActivity : BrowserTabActivity
```
nickrandolph, line 22: You will need this activity to support MSAL

UI/UnoSQLiteOneDriveInvoiceSample/Demo/Demo/Demo.Droid/Demo.Droid.csproj
```diff
@@ -1,148 +0,0 @@
...
-      <Version>4.24.0</Version>
-    </PackageReference>
-    <PackageReference Include="Microsoft.Identity.Client">
-      <Version>4.43.0</Version>
-    </PackageReference>
-    <PackageReference Include="Newtonsoft.Json">
-      <Version>13.0.1</Version>
-    </PackageReference>
-    <PackageReference Include="SQLiteNetExtensions">
-      <Version>2.1.0</Version>
```
nickrandolph, line 79: You need to double check which references are include in the original csproj files and make sure they're carried forward to the new projects

nickrandolph:
@aesteves900 there's a bunch of small bug fixes that need to be done but you'll need to make sure each project has the correct nuget references - this is a SQLite sample, so you should expect that all the projects reference sqlite.

UI/UnoSQLiteOneDriveInvoiceSample/Demo/Demo/Demo.Shared/App.xaml.cs
```diff
@@ -99,7 +99,7 @@ protected override void OnLaunched(LaunchActivatedEventArgs args)
             }
 
 #if !(NET5_0 && WINDOWS)
```
nickrandolph, line 101: @aesteves900 this needs to be changed to net5 or greater otherwise you end up with a crash on Windows

nickrandolph:
@aesteves900 I'm pushing this back to draft:
- Sample doesn't run on Windows (I haven't tested other targets)
- Please update WindowsApp Sdk to latest
- Please add Gtk/Wpf/Linux targets unless they're not supported

aesteves900:
@jeromelaban , on Windows we consistently got an error below: 
image


Do you have any clue?

jeromelaban:
Thanks. It's likely that this demo never worked on wasm, it's missing the `Uno.sqlite-wasm` package. Can you validate if it did without your changes?

aesteves900:
> Thanks. It's likely that this demo never worked on wasm, it's missing the `Uno.sqlite-wasm` package. Can you validate if it did without your changes?

Not working as well, there is the same error.

aesteves900:
The issue was created to solve the SQLLite problem on WASM platform.  https://github.com/unoplatform/Uno.Samples/issues/258

nickrandolph:
@aesteves900 more work required on this sample
- Windows doesn't run - I've pushed a fix that will allow it to run but still needs work to tidy up
- Does the OneDrive integration work?
- Why are these files not included in the Shared project
 
image

nickrandolph (approved):

PR #224, opened 2022-11-14
Create a Simple Calculator that targets .net 7

pictos:
I created a separated project to not mess up with the @eriklimakc's code.


nickrandolph:
@pictos is this still needed? or is the goal to combine this with the other simple calculator?

pictos:
@nickrandolph, no I believe this can be closed

PR #226, opened 2022-11-15
chore: Cupertino Sample - Update to net6

aesteves900:


nickrandolph:
@aesteves900 the windows target fails to run as it's failing to load resources. @kazo0 can you take a look at why resources aren't resolving on WinUI (we probably need WinUI sample apps in the Themes repo!)

nickrandolph (changes requested):
Blocking this PR until we get a fix for the cupertino styles for WinUI

nickrandolph:
@kazo0 @ajpinedam just tagging you on this PR. Once we have fixes for Themes, can you update this PR and make sure it's building and running on all platforms

kazo0:
Ah, I believe at least part of the issue is that, in WinUI, we can no longer use a string in the XAML as an ItemsSource like we are doing here:
https://github.com/unoplatform/Uno.Samples/blob/dev/AEST/190-update-to-net6-cupertino/UI/UnoCupertinoSample/UnoCupertinoSample.Shared/MainPage.xaml#L24

And the issue with the TextBox causing the crash has been fixed in the latest dev builds of Cupertino

/cc @aesteves900 @nickrandolph

kazo0:
Take a look at the Material PR, we changed it so ItemSource has a binding to a List in the code-behind

https://github.com/unoplatform/Uno.Samples/blob/dev/AEST/190-update-to-net6-material/UI/UnoMaterialSample/UnoMaterialSample.Shared/MainPage.xaml#L19

nickrandolph (approved):

PR #229, opened 2022-11-16
chore: TimeEntry Sample - Update to net6

aesteves900:


UI/TimeEntry/TimeEntryUno/TimeEntryUno.Windows/Properties/launchSettings.json
```diff
@@ -0,0 +1,7 @@
+{
```
nickrandolph, line 1: You need to make sure you add the pubxml files under PublishProfiles

nickrandolph (approved):

PR #234, opened 2022-11-17
Upgrading markup packages

eriklimakc:


nickrandolph:
@eriklimakc you should update to the latest stable for WinAppSdk as well as latest stable for Microsoft.Windows.SDK.NET.Ref in the Windows csproj file. This should fix your build error

pictos (approved):

PR #235, opened 2022-11-17
feat: use uno.themes markup constants

kazo0:


reference/SimpleCalculator/SimpleCalculator.Shared/MainPageMarkup.cs
```diff
@@ -80,15 +81,15 @@ private TextBlock Equation(MainModel.MainViewModel vm)
...
+            .Style(Theme.Styles.TextBlock.DisplaySmall);
 
         private TextBlock Result(MainModel.MainViewModel vm)
             => new TextBlock()
             .Text(() => vm.Calculator.Output)
             .AutoLayout(counterAlignment: AutoLayoutAlignment.End)
-            .Foreground(ThemeResource.Get<Brush>("OnBackgroundBrush"))
-            .Style(StaticResource.Get<Style>("DisplayLarge"));
+            .Foreground(Theme.Brushes.OnBackground.Default)
+            .Style(Theme.Styles.TextBlock.DisplayLarge);
```
nickrandolph, line 93: I think there are other brushes used in the KeyPad and KeyPadButton methods that should be able to use the constants

nickrandolph: @pictos or @eriklimakc can you pick this up and update the remaining methods with the theme constants

eriklimakc (approved):

reference/SimpleCalculator/SimpleCalculator.Shared/MainPageMarkup.cs
eriklimakc: We were instructed to leave it as it is for now @nickrandolph. But we may go back to it later.

PR #237, opened 2022-11-18
chore: StatusBarColor Sample - Update to net6

aesteves900:


aesteves900:
@nickrandolph , I mimic the sample behavior from the original sample and updated the packages.

nickrandolph:
@aesteves900 this sample is still broken. When you run on Android it should change the status bar color eg
image
Please review the original code and make sure the updated sample behaves the same.
Since this sample doesn't do anything for WASM and Windows targets, I suggest we drop those targets from the sample

UI/StatusBarThemeColor/StatusBarColorSample.Shared/App.xaml.cs
```diff
@@ -1,153 +1,172 @@
...
-					// When the navigation stack isn't restored navigate to the first page,
-					// configuring the new page by passing required information as a navigation
-					// parameter
-					rootFrame.Navigate(typeof(MainPage), e.Arguments);
-				}
-				// Ensure the current window is active
-				window.Activate();
-			}
-
-			ConfigureStatusBar();
```
nickrandolph, line 108: This link needs to be called in order for the statusbar color to be set

PR #244, opened 2022-11-22
chore: NativeFrameNav Sample - Update to net6

aesteves900:


aesteves900:
@nickrandolph , there is a reported issue on WinUI3 about SystemNavigationManager.

https://github.com/microsoft/WindowsAppSDK/issues/2398


It causes the crash on Windows:


image

Xiaoy312:
You can selectively disable the back-button supports (entire setup with `ConfigureNavigation`) on winui-desktop.
This sample is really created to showcase the native frame navigation on ios/droid, everything else is just bonus.

note: you should still keep as it is, for winui-ios/droid/anything-else as the back navigation depends on it in uno.

Xiaoy312 (approved):

PR #253, opened 2022-11-30
chore: ControlLibrary Sample - Update to net6

aesteves900:


nickrandolph:
@aesteves900 this sample is missing some of the uno targets (eg wpf, gtk, linux). Can you please add these to the solution
Also, it looks like the Wasm and Android targets aren't rendering the custom control. Can you please make sure you test the samples on all platforms.

aesteves900:
Hello @nickrandolph 

- Okay I will fix that
- Could we create an issue for that? According to Jerome we need to update each sample as-is.
- I fixing the issue, because the control rendering on Windows properly.

nickrandolph:
Closing this as this sample has already been updated now

PR #269, opened 2022-12-12
feat: Adding Pet Adopt UI sample

agaldanaw2:
These are some screenshots of the app running:

**Running PetAdopUI app:**

1. image
2. image
3. image


kazo0:
@mtmattei preliminary PR for the app

README.md
```diff
@@ -152,6 +152,10 @@ Note that for WebAssembly assets are downloaded on demand, as can be seen in the
 
 [Browse source](https://github.com/unoplatform/Uno.Samples/tree/master/UI/PackageResources)
 
+### Pet Adopt
+
```
nickrandolph, line 156: Can we add a description in here as to why this sample exists. Perhaps a link to the blog post

kazo0: Once the blog is published then we could update this description with the link

kazo0:
> Can we please add back all the other targets - where possible we should avoid samples that don't include the full set of uno targets

@nickrandolph this one is for a blog that looks to be specifically targeting mobile platforms based on the Figma designs. We could keep the other heads but the end result on a larger form factor would probably look pretty terrible.

Think maybe it's worth it to go back and get a design out for larger screens and have the blog cover both mobile and non-mobile versions of the UI?

nickrandolph:
@kazo0 I don't disagree with regards to design but unless there's an actual blocker that would prevent it running on a target, we should make sure we always include all targets for samples. Exception to this would be for samples that demonstrate a platform specific feature (statusbarcolor comes to mind)

nickrandolph (approved):

PR #282, opened 2022-12-26
feat: Update XamlBasics sample for alternating ListViewItems

lukeblevins:
FYI: @nickrandolph


nickrandolph:
@lukeblevins can we remove the Server project and add back all the supported Uno targets. This code applies equally to all uno targets, so the sample should include those targets

PR #300, opened 2023-01-16
feat: CardView sample migration to Uno

peterfoot:
New sample application to accompany blog post on migrating custom controls from Xamarin to Uno/WinUI XAML and code.

PR #303, opened 2023-02-01
feat: Adding Linux and WPF targets

nickrandolph:
Follow up on this PR (https://github.com/unoplatform/Uno.Samples/pull/300) to add linux and wpf targets. Also updating Uno references


weitzhandler:
Skia.WPF works well.

weitzhandler (approved):
Works on Skia.WPF

PR #304, opened 2023-02-08
Fixes the SplashScreen sample

pictos:
When we try to run the sample, it doesn't show the correct SplashScreen, so here I'm fixing it and making everyone happy.

- We need to update the docs for Android since google changed how it works on Android 12 and above.
- We need to update the docs for iOS, the VS doesn't edit `storyboard` files, we need to use XCode for it


UI/SplashScreenSample/SplashScreenSample.Mobile/iOS/LaunchScreen.storyboard
```diff
@@ -1,52 +1,44 @@
-﻿<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="11762" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" colorMatched="YES" initialViewController="5" launchScreen="YES">
+<?xml version="1.0" encoding="UTF-8"?>
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="21507" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" launchScreen="YES" colorMatched="YES" initialViewController="5">
```
nickrandolph, line 2: Did we need to modify the whole storyboard - aren't we just changing the image path?

UI/SplashScreenSample/SplashScreenSample.Wasm/WasmScripts/AppManifest.js
```diff
@@ -1,6 +1,6 @@
 ﻿var UnoAppManifest = {
 
-    splashScreenImage: "Assets/SplashScreen.png",
+    splashScreenImage: "Assets/SplashScreen.scale-200.png",
```
nickrandolph, line 3: For web I think the 100x is probably the better option.

pictos: I just followed our doc
https://platform.uno/docs/articles/splash-screen.html?tabs=tabid-vswin#step-5---webassembly

UI/SplashScreenSample/SplashScreenSample.Mobile/iOS/LaunchScreen.storyboard
pictos: I opened on xcode, so it updated the rest, I presume... I can revert it and edit the `xml` file if you want

nickrandolph (approved):

PR #318, opened 2023-03-23
Add MVUX how-to samples

weitzhandler:
This PR includes several MVUX how-tos that are used in the MVUX docs.


UI/MvuxHowTos/PeopleApp/PeopleApp/PeopleService.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        await Task.Delay(TimeSpan.FromSeconds(2), ct);
+
+        var people = new Person[]
+        {
+            new Person(FirstName: "Master", LastName: "Yoda"),
+            new Person(FirstName: "Darth", LastName: "Vader")
+        };
+
+        return people.ToImmutableList();
```
roubachof, line 17: At last, someone using real immutable collections 👏

roubachof (commented):
Nice and clean work!
I'm just puzzled about the name Model designating the "view model" layer.
Models are normally assimilable to entities (in a ddd terminology), they are basically data.
Here models are just Controllers really: they route the data back and forth between the UI and the Services.
MVU is not a complete layered architecture it's more like flux immutable updates + declarative UI. So I think it could be ok to name Controller instead of Model the class who acts as the glue between UI and Domaim.

weitzhandler:
@roubachof 
Thanks for your comment!
In MVUX, Model takes over the place of the MVVM ViewModel. And they do hold entities.
MVUX is not fundamentalistically MVU. Though regarding the terminology, it's FT, @dr1rrb and @nickrandolph who decide.

nickrandolph:
@roubachof thanks for the feedback - naming is hard but we'd like to stay away from controller because it implies some level of control over the view layer when in fact there's no reference to the view from the model layer. One of the important aspects of mvux is that the model is immutable and yet we're still able to fully leverage the binding engine for one and two way binding, including command binding.

PR #343, opened 2023-06-05
ChatGPT Demo

Onotseike:
This PR adds sample code to showcase how to consume Open AI's Chat GPT APIs in an Uno Platform application


MartinZikmund (approved):

MartinZikmund:
@Guidemarcus API key in the code is valid? Ideally you should remove it from the repository, otherwise people will be using it and you will get charged. If it is a real key, please also remove it from your account

Onotseike:
> @Guidemarcus API key in the code is valid? Ideally you should remove it from the repository, otherwise people will be using it and you will get charged. If it is a real key, please also remove it from your account

I have removed it from my account

Issue #416, opened 2023-07-28
[Commerce] The Description does not contain what features it demonstrates


Marc-Antoine-Soucy:
The description of the the Commerce app is this: "The The Commercial App Sample with Cart, products and so on." Which tells us nothing about what it demonstrates.

Marc-Antoine-Soucy:
My proposition: The commerce app, is a sample app demonstrating the use of ListFeed pagination, Feedviews, and other features of Uno.Extensions can be used to make an app with a cart, products and so on.

PR #423, opened 2023-07-28
fix: commerce app description

Marc-Antoine-Soucy:
#416 

Update description so that it includes what features are being used.


Marc-Antoine-Soucy:
We could have a more comprehensive description, if we want.

nickrandolph (approved):

Issue #415, opened 2023-07-27
[Commerce] Pages are empty


Marc-Antoine-Soucy:
Issue: Right now the deal, product, and cart page are empty

In an older version of the app, the deal page had at least some products:
![image](https://github.com/unoplatform/Uno.Samples/assets/90481654/ef8bd948-1a0e-4e6e-b79c-505f00798d91)
Right now, it's just empty.  And all the other pages are also empty. This seems because the feed doesnt receive any data 
![image](https://github.com/unoplatform/Uno.Samples/assets/90481654/d9047168-d661-4602-8afc-3dd4ad2f671a).

Marc-Antoine-Soucy:
It seems to be because IStorage (Uno.Extensions.Storage.IStorage) doesnt return anything, which is probably related to this warning we get in the output: 
```
Uno.Extensions.Storage.FileStorage: Warning: OpenPackageFileAsync - File 'products.json' does not exist in package
Uno.Extensions.Storage.FileStorage: Warning: OpenPackageFileAsync - File 'products.json' does not exist in package
```
It seems to say it's searching for the file in the package, when it's in the app? 
![image](https://github.com/unoplatform/Uno.Samples/assets/90481654/ee4b829b-d500-4943-bb11-268722165e13)

But I'm not quite sure what to do about this. I didnt find documentation about how to use the package.

ttustonic:
It seems that files 'products.json' and 'reviews.json' should have 'Copy to Output Directory' property set.

ttustonic:
It works for me on Android (didn't try Wasm), after I uninstalled old Commerce from the emulator.

Marc-Antoine-Soucy:
Hm, this is a little weird. Did you do any other change or only ( Copy to Output Directory => Copy always)? It could also be related to the Android version.

ttustonic:
Hmm, even weirder... it works when I set 'Copy if newer', but not with 'Copy always' !?
I didn't change anything else (I think), just Clean, Rebuild and uninstall old version.
My Android emulator is Pixel 5 Android 12.0 - API 31

Marc-Antoine-Soucy:
Thanks for the help, @ttustonic, copy if newer appears to work for all testable platforms.

PR #426, opened 2023-08-01
fix: empty pages

Marc-Antoine-Soucy:
This PR ensures 'products.json' and 'reviews.json' are copied to the output directory at runtime, resolving a file access issue that made the pages empty.

#415


nickrandolph:
If these files are marked as Content then they should automatically be copied to the output folder. Let's hold off merging this for further investigation

nickrandolph (approved):

PR #480, opened 2023-08-08
Sample code for UnoExpandableParagraph

Onotseike:
This sample code shows how to create a custom expandable paragraph control that also has a header text and footer button (customisations include)


nickrandolph:
@Onotseike would you mind removing the .DS_Store files and if possible adding the Gtk target platform

Onotseike:
> @Onotseike would you mind removing the .DS_Store files and if possible adding the Gtk target platform

Done. I did not add the gtk since I don't have a way to test it out and I don't want to add a platform I haven't tested to the code.

PR #494, opened 2023-08-09
fix: can deploy on windows

### LINKED ISSUE #493: [SkottieSample] Windows crashes on start
Exception thrown: 'System.Runtime.InteropServices.COMException' in WinRT.Runtime.dll
Exception thrown at 0x00007FFD4427CD29 (KernelBase.dll) in SkottieSample.Windows.exe: WinRT originate error - 0x80070490 : 'Element not found. (0x80070490)'.
onecore\com\combase\winrt\error\restrictederror.cpp(1016)\combase.dll!00007FFD46165EC3: (caller: 00007FFD4612CA9D) ReturnHr(1) tid(8eac) 8007007E The specified module could not be found.
Microsoft.ui.xaml.dll!00007FFCB099380E: 80070490 - Element not found.

Marc-Antoine-Soucy:
Closes #493

### DIFF
--- a/UI/SkottieSample/SkottieSample/SkottieSample.Shared/MainPage.xaml.cs
+++ b/UI/SkottieSample/SkottieSample/SkottieSample.Shared/MainPage.xaml.cs
@@ -16,8 +16,6 @@ using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices.WindowsRuntime;
 using System.Threading.Tasks;
-using Windows.Foundation;
-using Windows.Foundation.Collections;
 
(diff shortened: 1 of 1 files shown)


Marc-Antoine-Soucy:
@nickrandolph  I think just deleting that line should be fine, since it doesnt do anything important apart from crashing windows. But, if it's better to replace it by an equivalent, I can try. (Microsoft.UI.ViewManagement.ApplicationView.GetForCurrentView().Title = "Skottie WebAssembly"; Does not work.

PR #509, opened 2023-08-16
fix: Updating to net7 and dropping xamarin

nickrandolph:


UI/HomeIdeasUI/Directory.Packages.props
```diff
@@ -0,0 +1,25 @@
+<Project ToolsVersion="15.0">
+  <ItemGroup>
```
Youssef1313, line 2: We are inconsistent whether we use two spaces or a tab for XML props/targets/csproj files.

This doesn't block merging this PR, but something we should unify.

Youssef1313 (approved):

PR #557, opened 2023-09-07
feat: Adding telerik app

nickrandolph:


UI/MauiEmbedding/TelerikApp/TelerikApp/Presentation/MainViewModel.cs
```diff
@@ -0,0 +1,47 @@
...
+
+	public ICommand Counter { get; }
+
+	private async Task GoToSecondView()
+	{
+		await _navigator.NavigateViewModelAsync<SecondViewModel>(this, data: new Entity(Name!));
+	}
+
+
+	private void OnCount()
```
pictos, line 39: Not sure if this was intentional, but you don't need to declare the `ICommand` interface, you can use the [RelayCommand] attribute on methods that should be Commands and the toolkit will generate the Commands for you

pictos (approved):

PR #597, opened 2023-11-03
chore: remove api key

kazo0:
The API Key that was there has surely already been scraped by bots and harvested for all its requests as I am just seeing the Error state now if I run this app as is. We should leave this as a placeholder


Xiaoy312 (approved):
perhaps we should leave a note somewhere to remind the user or the future us, that this needs to be replaced

PR #600, opened 2023-11-07
fix: Image source

weitzhandler:


reference/TubePlayer/TubePlayer/Presentation/MainPage.cs
```diff
@@ -52,7 +52,7 @@ public MainPage()
                                     .Children
                                     (
                                         new Image()
-                                            .Source(new SvgImageSource(new Uri("ms-appx:///TubePlayer/Assets/navigation_bar")))
+                                            .Source("ms-appx:///TubePlayer/Assets/navigation_bar.png")
```
weitzhandler, line 55: Do we need to specify the extension?

nickrandolph: Yes

nickrandolph (approved):

PR #611, opened 2023-12-11
Update code on mvvm samples for SimpleCalc

pictos:
I updated the `MainViewModel.IsDark` code to use correctly the features of MVVM Toolkit.


reference/SimpleCalc/MVVM-CSharp/SimpleCalculator/MainViewModel.cs
```diff
@@ -1,24 +1,14 @@
-﻿using CommunityToolkit.Mvvm.ComponentModel;
-using CommunityToolkit.Mvvm.Input;
-
 namespace SimpleCalculator;
 
 public partial class MainViewModel : ObservableObject
 {
-	private readonly IThemeService _themeService;
+    private readonly IThemeService _themeService;
```
nickrandolph, line 5: Can you make sure that spacing/tabs is consistent with rest of the code

pictos: @nickrandolph used tabs. But the `editorconfig` used by the template is set to space and not tabs. I found that's the default on our template.
image

nickrandolph: happy for the code to change to tabs but we should apply that across all code, not just on one file

pictos:
@nickrandolph I believe it's good to go now

nickrandolph (approved):

PR #624, opened 2024-01-09
chore: update uno

Robert-Louis:
Update uno fixing the layout trigger arrange phase multiple time
https://github.com/unoplatform/uno/issues/14373

### DIFF
--- a/reference/TubePlayer/Directory.Packages.props
+++ b/reference/TubePlayer/Directory.Packages.props
@@ -29,20 +29,20 @@
     <PackageVersion Include="Uno.Wasm.Bootstrap" Version="8.0.4" />
     <PackageVersion Include="Uno.Wasm.Bootstrap.DevServer" Version="8.0.4" />
     <PackageVersion Include="Uno.Wasm.Bootstrap.Server" Version="7.0.31" />
-    <PackageVersion Include="Uno.WinUI" Version="5.0.64" />
-    <PackageVersion Include="Uno.WinUI.Lottie" Version="5.0.64" />
+    <PackageVersion Include="Uno.WinUI" Version="5.1.80" />
+    <PackageVersion Include="Uno.WinUI.Lottie" Version="5.1.80" />
     <PackageVersion Include="Uno.WinUI.Markup" Version="5.0.13" />
     <PackageVersion Include="Uno.Toolkit.WinUI.Markup" Version="5.0.17" />
     <PackageVersion Include="Uno.Extensions.Navigation.WinUI.Markup" Version="3.0.11" />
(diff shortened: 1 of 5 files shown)


reference/TubePlayer/Directory.Packages.props
```diff
@@ -29,7 +29,7 @@
     <PackageVersion Include="Uno.Wasm.Bootstrap" Version="8.0.4" />
     <PackageVersion Include="Uno.Wasm.Bootstrap.DevServer" Version="8.0.4" />
     <PackageVersion Include="Uno.Wasm.Bootstrap.Server" Version="7.0.31" />
-    <PackageVersion Include="Uno.WinUI" Version="5.0.64" />
+    <PackageVersion Include="Uno.WinUI" Version="5.1.0-dev.1268" />
```
agneszitte, line 32: @nickrandolph, @jeromelaban 
I think the other Uno.WinUI versions should match the same version as well (Uno.WinUI....). 
Also, the fix (https://github.com/unoplatform/uno/pull/14902) should be backported no ? 
So we can use a proper stable version for the TubePlayer sample app but also the same version update will be needed for the SimpleCalc sample app (cf. screenshot and details here: https://github.com/unoplatform/uno.toolkit.ui-private/issues/7)

agneszitte (changes requested):

reference/TubePlayer/Directory.Packages.props
Youssef1313: We shouldn't be using dev/prerelease versions. We should either wait for 5.1 stable or backport.

jeromelaban: Indeed, let's wait for the 5.1 release. Let's not backport either for now, it's a very significant change.

nickrandolph (changes requested):
Please don't use dev versions in samples repository

agneszitte:
> Please don't use dev versions in samples repository

@Robert-Louis  Now that 5.1 stable is available can you update your PR please so we can have the update for 5.1 stable for **TubePlayer** but also for **SimpleCalc** (cf. screenshots in https://github.com/unoplatform/Uno.Samples/pull/624)

Thanks in advance

eriklimakc (approved):
Tested all 4 versions of SimpleCalc and the TubePlayer app. No regression found, the apps are working as expected.

agneszitte (approved):

agneszitte:
> Tested all 4 versions of SimpleCalc and the TubePlayer app. No regression found, the apps are working as expected.

Thank you so much for the quick sanity check @eriklimakc, really appreciated!

PR #627, opened 2024-01-18
chore(SkiaSharpTest): Migrating WpfHost in SkiaSharpTest

lindexi:
Fix the running error as:

> If you are seeing this, make sure to follow the "Migrating WpfHost" section of Migrating from previous releases article in the Uno Platform documentation at https://aka.platform.uno/uno5-wpfhost-migration. WpfHost is used at the application level instead of window level starting Uno Platform 5.0.


lindexi:
@jeromelaban Can I ask any update here?

jeromelaban:
@lindexi Apologies for the delay. we'll likely update that sample to use single project, which won't use that structure anymore.

Let's try a rebuild to be sure.

lindexi:
@jeromelaban What should I do? Thank you.

jeromelaban (approved):
The build looks good, let's merge as is :) @Kunal22shah may change the structure later one as it gets migrated to single-project.

PR #630, opened 2024-01-24
chore: ChatGPT refactor

eriklimakc:


UI/ChatGPT/ChatGPT/ChatGPT/Presentation/MainModel.cs
```diff
@@ -47,18 +47,16 @@ private async ValueTask Ask(string prompt, CancellationToken ct)
...
-		var request = new ChatRequest(history);
+		//Create ChatRequest record with messages list
+		var request = new ChatRequest(requestContent);
 
 		var response = await _chatService.AskAsync(request);
 
 		//Update loading message with AI response as Message record
-		await Update(message, response, ct);
+		//Finds the message with same id and upadtes the instance
+		await Messages.UpdateAsync(message.Update(response), ct);
```
nickrandolph, line 59: Does this add to the list of Messages, or amend an existing message - if the latter, feels very not-immutable :-)

UI/ChatGPT/ChatGPT/ChatGPT/Presentation/Message.cs
```diff
@@ -9,4 +11,13 @@ public Message(string userMessage)
 		: this(Guid.NewGuid(), Source.User, Status.Value, userMessage)
 	{
 	}
+
+	public Message Update(ChatResponse response)
```
nickrandolph, line 15: I'm not sure I'm a fan of using Update here as it implies changing current instance. Perhaps "With" to be consistent with how records are updated
eg msg.With(response)
cc @francoistanguay thoughts?

francoistanguay: Fait. Maybe "With" as a With(reponse) : Message...

We'll see those patterns evolve... in ToDo... we had Todo.ToggleIsCompleted():ToDo as itwas modifying only one field and not accepting a param...

I dont think a method named .Response(reponse) would make much sense but maybe?

It's just a sample but I can see why Update has more mutable meaning and could be congusing.

UI/ChatGPT/ChatGPT/ChatGPT/Presentation/MainModel.cs
eriklimakc: It looks for an item with same `Id` and if the item is found, the item is removed and the updated (new) item is added.

[Here is the `UpdateAsync`](https://github.com/unoplatform/uno.extensions/blob/main/src/Uno.Extensions.Reactive/Core/ListState.Extensions.cs#L199) code in Uno.Extensions.Reactive.

UI/ChatGPT/ChatGPT/ChatGPT/Presentation/Message.cs
eriklimakc: Maybe `message.WithResponse(response)` would be more concise?

francoistanguay: Sure. That's fine. As it was the only "With" or that we could even have overloads... I was fine with .With simply... But if you prefer yours, go for it.

UI/ChatGPT/ChatGPT/ChatGPT/Presentation/MainModel.cs
francoistanguay: ListState is "Mutable" conceptually... Just like State.... This is why it exists in the first place....

It's all good

agneszitte (approved):

PR #660, opened 2024-04-16
feat: Adding dynamic data sample

nickrandolph:
This sample demonstrates how data can be loaded from two different sources based on the platform type. In this case when run as Desktop the data is loaded using the LocalData service (in a real app this could be EF talking directly to a SQL db), otherwise data is loaded via the Rest api.


nickrandolph:
cc @jeromelaban @dansiegel @dr1rrb let me know what you think about the structure of this sample app

PR #684, opened 2024-05-13
chore: Update PongWars GamePage 

morning4coffe-dev:


UI/PongWars/src/UnoPongWars/Presentation/GamePage.cs
```diff
@@ -18,52 +18,52 @@ public GamePage()
...
-                                        .Width(10)
-                                        .Height(10)
-                                        .Fill(x => x.Binding(() => cell)
-                                            .Convert(cell => new SolidColorBrush(PlayerColor(cell))))
-                                        .Visibility(x => x.Binding(() => cell)
-                                            .Convert(cell => cell.HasBall ? Visibility.Visible : Visibility.Collapsed))))),
-                        new TextBlock()
-                            .Text(x => x.Binding(() => vm.Score))
+                    .Margin(20)
+                    .RowDefinitions("*,auto,auto")
```
agneszitte, line 22:
```suggestion
                    .RowDefinitions("*,Auto,Auto")
```

UI/PongWars/src/UnoPongWars/Presentation/GamePage.cs
```diff
@@ -18,52 +18,52 @@ public GamePage()
...
+                                    .Text(() => vm.Score)
+                                    .HorizontalAlignment(HorizontalAlignment.Center)
+                                    .Style(Theme.TextBlock.Styles.HeadlineSmall)
+                                    .Margin(10)
+                                    .Grid(row: 1),
+                                new Slider()
+                                    .MaxWidth(400)
+                                    .Margin(16)
+                                    .Maximum(1000)
 #if __ANDROID__
```
agneszitte, line 47: @dr1rrb, @morning4coffe-dev I think we would need to add a comment here to give a bit more info on why we have a `#if __ANDROID__` here

agneszitte:
@morning4coffe-dev I tested your changes on all platforms and it is working well in general.
But while doing a bit more tests with the slider, I noticed this issue below.
It is not related to your current changes in this PR as I am able to reproduce it with the current version on `master` as well.
Before opening an issue, I would love your feedback @dr1rrb, please.

**[I can only reproduce it for Android and iOS]**
So basically to repro the issue you need to push to the **maximum speed** then try to reduce the speed and you can notice that the speed is not updating properly.
- **Android**
PongWars-android-speedissue
- **iOS**
PongWars-ios-speedissue

morning4coffe-dev:
Thanks @agneszitte, interesting, I was able to repro it and from my quick investigations the Speed property is updated correctly, so the issue must be that the Delay is somehow not updated properly which is very weird. I will try to investigate more, in the mean time, can we merge this PR once we adjust with your suggestions and make an Issue from this? (as it is not related to this PRs updates)

agneszitte:
> Thanks @agneszitte, interesting, I was able to repro it and from my quick investigations the Speed property is updated correctly, so the issue must be that the Delay is somehow not updated properly which is very weird. I will try to investigate more, in the mean time, can we merge this PR once we adjust with your suggestions and make an Issue from this? (as it is not related to this PRs updates)

@morning4coffe-dev you can go ahead with the suggestions and I have opened a separate issue for the Delay: https://github.com/unoplatform/Uno.Samples/issues/686

dr1rrb:
For the delay "issue", I think we could just ignore it. It's a limitation of the "MVUX as a game engine" model ... which is obviously not a real concern.

agneszitte (approved):

Issue #727, opened 2024-05-27
[All platforms/TubePlayer] TextBox Placeholder is cut off


rajamatt:
![image](https://github.com/unoplatform/Uno.Samples/assets/70542961/97c3a197-f11b-439a-bb5a-dcab61d23172)

rajamatt:
Will try to repro this in a minimal sample app.

rajamatt:
After testing in a new uno app, here's what I found:

For a default Material theme TextBox, the placeholder text is cut off when the height is set to anything lower than 62.

```csharp
new TextBox()
    .Height(61) 
    .PlaceholderText("Placeholder")
```

![image](https://github.com/unoplatform/Uno.Samples/assets/70542961/dfc1d737-d737-40d7-8f8b-8c526556f782)

```csharp
new TextBox()
    .Height(62)
    .PlaceholderText("Placeholder")
```

![image](https://github.com/unoplatform/Uno.Samples/assets/70542961/661d7161-4b7b-4800-a7ad-def896bb39f2)

rajamatt:
@kazo0 is this a known thing with textboxes?

kazo0:
@rajamatt yes I believe it's because we are hardcoding a height in the code there. We should keep it at the default height for now

agneszitte:
> Let's just remove the Height for now, we can revisit once [unoplatform/Uno.Themes#1014](https://github.com/unoplatform/Uno.Themes/pull/1014) is merged now that it's unblocked

@kazo0 I am okay with removing the fixed height, it is making more sense also in my opinion. But @NVLudwig the generated code from the figma file would need to follow as well in addition to @rajamatt updating the workshop-related code for the documentation.

rajamatt:
@agneszitte I updated the workshop docs in this PR: https://github.com/unoplatform/workshops/pull/164
And removed the height in the singleproject PR: https://github.com/unoplatform/Uno.Samples/pull/728

NVLudwig:
**A brief history of TextBox**
1. Tube Player design was made to repro an external design. 
2. When we started we identified a limitation of Uno TextBox: it did not allow a textBox to not have a floating label, it have a hard coded min height and did not allow trailing Icon buttons.
3. It was agreed we would create code for a TextBox style to allow having not floating label and trailing icon buttons and no min height.
[...]

Issue #725, opened 2024-05-27
[Desktop] Videos don't play with latest packages


rajamatt:
Noticed for TubePlayer when updating to latest packages for Single Project migration.

- Desktop: no video controls on Desktop (clicking on video does nothing either)

Desktop:
![image](https://github.com/unoplatform/Uno.Samples/assets/70542961/a1f9e3fe-553b-4617-b69b-8ad6030bae86)

rajamatt:
I will try to create a small repro app for this.

rajamatt:
Just started investigating this and I found that Skia Desktop doesn't support MediaPlayerElement? https://platform.uno/docs/articles/controls/MediaPlayerElement.html

jeromelaban:
@rajamatt this is expected at this time, yes.

PR #728, opened 2024-05-27
chore: singleproject migration TubePlayer app

### LINKED ISSUE #726: [TubePlayer] `PublishedAt` date logic is missing a Math.Abs or invert subtraction
TubePlayer YouTube videos published date is always "less than a minute ago" because of the logic

We should change the logic to display the real date, like this 6 months ago:
image

We need to change it in the tutorial too: https://platform.uno/docs/articles/external/workshops/tube-player/modules/03-Connect-UI-with-mock-data/README.html#create-entities

rajamatt:
There are two issues that seem to be related to bumping the packages for single project:

- [x] https://github.com/unoplatform/Uno.Samples/issues/725
- [x] https://github.com/unoplatform/Uno.Samples/issues/727

Includes fix for: closes #726


rajamatt:
@agneszitte seems like workaround for https://github.com/unoplatform/Uno.Samples/issues/727  is to just use default height, should I remove the `.Height(40)` setter for the search TextBox? It looks like this on all platforms currently: 
image

agneszitte (approved):

PR #754, opened 2024-06-11
feat: ChatGPT - SingleProject

eriklimakc:


UI/ChatGPT/src/ChatGPT/Services/ChatService.cs
```diff
@@ -95,4 +95,4 @@ private ChatCompletionCreateRequest ToCompletionRequest(ChatRequest request)
 			Model = Models.Gpt_3_5_Turbo
```
carldebilly, line 95: We may want to upgrade to a newer, better, faster model ;-)

carldebilly (approved):

UI/ChatGPT/src/ChatGPT/Services/ChatService.cs
agneszitte: @carldebilly not in this PR, we are keeping things the same as before to migrate to single project. We can make improvements in following PRs

agneszitte (approved):

PR #720, opened 2024-05-24
fix: ChatSignalR - stop hang on WASM

mikeirvingweb:
rectified an issue with `Delay` which would cause a permanent hang in Web Assembly.  

perhaps a temporary fix, until `threading` in .NET 9, as alluded to by @jeromelaban [here](https://github.com/unoplatform/uno/issues/13370)  

for the ChatSignalR sample:
 - closes #417 
 - closes #430 
 
 - *should be the fix that* closes #118 
   - no connection issues once fixed.. the observed issue was likely the hang
   - the only connection changes I have ever had to make are to change the WASM port from 5000 to 5001- likely no relation

PR #762, opened 2024-06-19
chore: improvement to wasm delay fix from #720

mikeirvingweb:
@jeromelaban I noticed that the #720 fix, and your suggested simplification had both been merged, creating a pointless nested if clause.

anyway, it should now be as desired 👍


jeromelaban (approved):
Thanks for keeping track!

Issue #774, opened 2024-07-05
[ToDo][Task List] Need to Toggle Twice to Favorite/Unfavorite a Task from Task List


agneszitte:
## Current behavior

You need to toggle the button twice to favorite or unfavorite a task from the task list.

https://github.com/unoplatform/Uno.Samples/assets/16295702/2f9619a8-3bef-4778-8a31-5dacd15fdee9

## Expected behavior

Toggling the button once should favorite or unfavorite the task from the task list.

## How to reproduce it (as minimally and precisely as possible)

- Open the application.
- Click on the "Get Started" button on the Welcome Page.
- Select any task from the first task list.
- Click the favorite icon in the task list to favorite the task.
- Observe that you need to toggle the button twice to favorite the task.
- Click the favorite icon again to unfavorite the task.
- Observe that you need to toggle the button twice to unfavorite the task.

## Environment

Nuget Package:
- Uno.Sdk: 5.2.161
- UnoExtensionsVersion: 4.1.28
- UnoToolkitVersion: 6.0.24
- UnoThemesVersion: 5.0.13

Affected platform(s):

- [X] All

PR #784, opened 2024-07-22
Fix Toggle Task Twice

Kunal22shah:
closes: #774

### DIFF
--- a/reference/ToDo/src/ToDo/Views/TaskListPage.xaml
+++ b/reference/ToDo/src/ToDo/Views/TaskListPage.xaml
@@ -113,7 +113,7 @@
                       <utu:AutoLayout PrimaryAxisAlignment="End"
                                       Orientation="Horizontal">
                         <ToggleButton utu:AutoLayout.CounterAlignment="Start"
-                                      IsChecked="{Binding IsImportant}"
+                                      IsChecked="{Binding IsImportant, Mode=TwoWay}"
                                       Command="{utu:AncestorBinding Path=DataContext.ToggleIsImportant, AncestorType=uer:FeedView}"
                                       CommandParameter="{Binding}"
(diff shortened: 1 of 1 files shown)


reference/ToDo/src/ToDo/Views/TaskListPage.xaml
```diff
@@ -113,7 +113,7 @@
                       <utu:AutoLayout PrimaryAxisAlignment="End"
                                       Orientation="Horizontal">
                         <ToggleButton utu:AutoLayout.CounterAlignment="Start"
-                                      IsChecked="{Binding IsImportant}"
+                                      IsChecked="{Binding IsImportant, Mode=TwoWay}"
```
kazo0, line 116: We shouldn't need this TwoWay binding, it should be properly coming downstream from the resulting update that is done as part of the ToggleIsImportant Command. It looks like we are subscribing to the right messages being sent by the ITaskService.

If you look [here](https://github.com/unoplatform/Uno.Samples/blob/d82132390ce1e59f9ab7d66a16fc5ac95651f455/reference/ToDo/src/ToDo/Business/TaskService.cs#L32) this is what is being called when we update a ToDoTask. You'll notice that we are sending a `EntityMessage<ToDoTask>`. So we should be observing the messenger for our `IListState<ToDoTask> Tasks` property in the view model. I see we are doing this and combining it with the Entity [here](https://github.com/unoplatform/Uno.Samples/blob/f97fccb502d3c890b72c94a46a6528742b9e5bb6/reference/ToDo/src/ToDo/Presentation/TaskListViewModel.cs#L28)

I'd be curious if this gets fixed if we were to remove the added TwoWay binding and add something like `messenger.Observe(Tasks, t => t.Id);` to the list of Observe calls in the constructor of our viewmodel.

kazo0 (approved):

PR #796, opened 2024-08-02
chore: Move `BluetoothExplorer` to SingleProject Uno.Sdk 5.3

morning4coffe-dev:
The old project wasn't building for me at all. The Windows version seems to work well now

image

NOTE: I have tested only on Windows for now, I will test Android


UI/BluetoothExplorer/BluetoothExplorer/BluetoothExplorer/Platforms/Android/AndroidManifest.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+  <application android:allowBackup="true" android:supportsRtl="true"></application>
```
Youssef1313, line 3: This is missing bluetooth permissions that used to exist in the old AndroidManifest.xml

UI/BluetoothExplorer/BluetoothExplorer/BluetoothExplorer/Platforms/iOS/Info.plist
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Youssef1313, line 1: This is probably also missing bluetooth permissions.

morning4coffe-dev:
I tested this on Android and Windows and I am not able to properly get it working on both, the `GetPrimaryServicesAsync` doesn't seem to return anything on my side when connecting to any device. I get the same result when validating this with master.

Issue #785, opened 2024-07-23
Update samples to latest Uno.Sdk 5.3


agneszitte:
Now that version 5.3 has been released, we need to update the samples to the latest stable Uno.Sdk 5.3.

Here's the markdown table with the specified columns:

✅: DONE
👀: In review
⌛: In progress
🚧: Needs re-work (ci errors, additional review comments,...)
🅱️: Blocked (+ issue and comment with the reason)
❌: Cannot be updated (+ comment with the reason)
⏸️: On hold (Not a priority to update older samples at the moment as it needs to migrate to the single structure at the same time)
📋: To-Do

| NAME                                   | ASSIGNEE     | STATUS | RELATED PR(s) | COMMENT |
|----------------------------------------|--------------|--------|---------------|---------|
| **Commerce App**                           | @Kunal22shah | 🚧    | https://github.com/unoplatform/Uno.Samples/pull/790 |         |
| **Counter**                                | @Kunal22shah | ✅   | https://github.com/unoplatform/Uno.Samples/pull/789 |         |
| **SimpleCalc App**                         | @Kunal22shah | ✅     | https://github.com/unoplatform/Uno.Samples/pull/786 |         |
| **ToDo App**                              | @Kunal22shah | 🚧   | https://github.com/unoplatform/Uno.Samples/pull/788 |         |
| **TubePlayer App**                         | @Kunal22shah | 🚧    | https://github.com/unoplatform/Uno.Samples/pull/787 |         |
| Advanced XBind                         |             @morning4coffe-dev        |    👀       |      #797          |         |
| Android Custom Camera                  |             @morning4coffe-dev        |    ✅       |        #806        |         |
| Authentication with OpenID Connect (OIDC) |         @morning4coffe-dev        |    🅱️/⏸️       |               | (Let's put it on hold for now) The sample doesn't seem to work outside of WASM. Fails on WebAuthenticationBroker.GetCurrentApplicationCallbackUri(); not implemented? We could update this with the newer Extension OIDC option?  |
| Auto-Suggest                           |       @morning4coffe-dev       |    ✅    |       #798         |         |
| Benchmark                              |       @morning4coffe-dev       |     ✅   |       #811         |         |
| BluetoothExplorer                      |      @morning4coffe-dev        |    🅱️/⏸️    |      #796          |    https://github.com/unoplatform/Uno.Samples/pull/796#issuecomment-2281107518     |
[...]

PR #804, opened 2024-08-09
chore: update CountryDataSample to singleproject and 5.3

rajamatt:
Updates CountryDataSample to singleproject structure and uno 5.3.
#785


UI/CountryDataSample/src/CountryDataSample/Platforms/Desktop/Program.cs
rajamatt, line 1: Sample just opens as a black screen on Desktop and Wasm with no relevant console output. Will investigate further later.

Issue #828, opened 2024-09-27
Update samples to latest Uno.Sdk 5.4


agneszitte:
Now that Uno.Sdk version **5.4** has been released, we need to update the samples to the **latest stable Uno.Sdk 5.4**.

**Migration notes for 5.4 can be found here: [Uno Platform 5.4 Migration Guide](https://github.com/unoplatform/uno/blob/master/doc/articles/migrating-from-previous-releases.md#uno-platform-54)**

> [!IMPORTANT]  
> **Samples with a status of ⏸️ (On hold), which require migration to the single project structure, are not a priority to be updated or worked on at the moment. The complete list of samples is included in the matrix below to provide an overview of all current samples in the Uno.Samples repo and their statuses between release updates.**

## Steps to follow for each sample you are updating:
- Assign your name in the table below before starting work to ensure no duplication of effort.
- If override versions are still present in `Directory.Build.props`, remove them. 
  [They are no longer necessary with Uno.Sdk 5.3 and above.](https://github.com/unoplatform/uno/blob/master/doc/articles/migrating-from-previous-releases.md#single-project-updates)
- Perform a quick sanity check by testing the sample app on all supported platforms.
- If a README is missing at the root of the sample, please create one. 
  Use the [WindowingSample README.md](https://github.com/unoplatform/Uno.Samples/tree/master/UI/WindowingSamples) or the [Counter README.md](https://github.com/unoplatform/Uno.Samples/tree/master/reference/Counter) as references for format and folder organization.
- If the sample is not listed in [samples.md](https://github.com/unoplatform/Uno.Samples/blob/master/doc/samples.md), ensure it is added.
- Adjust the status and the last current Uno.Sdk version used by the sample in the table below when changes are merged

---

### Status Legend:

- **📋**: To-Do
- **⌛**: In progress
- **👀**: In review
- **🚧**: Needs re-work (CI errors, additional review comments, etc.)
- **🅱️**: Blocked (Issue and comment with the reason)
- **⏸️**: On hold (Not a priority to update older samples at the moment as they need to migrate to the single structure first)
- **✅**: DONE

### ReadMe.md/Samples.md Legend:
- **❌**: MISSING (If "Descr" is mentioned, it means that the sample is listed but missing the related description)
- **✅**: AVAILABLE

[...]

PR #843, opened 2024-10-04
chore: Update MediaGallery to Uno.Sdk 5.4

eriklimakc:
Issue: https://github.com/unoplatform/Uno.Samples/issues/828

Update Sdk to 5.4.5
Some files are using tab and others space for indentation. Changes for all to use space.


eriklimakc:
@agneszitte @kazo0 Should we remove MacOS and Windows from the Platforms folder/launchSettings.json (only Windows), since this app is Android and iOS only?

https://github.com/unoplatform/Uno.Samples/blob/19d039de6cdfa48d343ac0913f591d99481c5b50/UI/MediaGallery/src/MediaGallerySample/MediaGallery/MediaFileType.cs#L1

vatsashah45 (approved):

UI/MediaGallery/src/global.json
```diff
@@ -1,6 +1,6 @@
 {
   // To update the version of Uno please update the version of the Uno.Sdk here. See https://aka.platform.uno/upgrade-uno-packages for more information.
   "msbuild-sdks": {
-    "Uno.Sdk": "5.3.99"
+    "Uno.Sdk": "5.4.5"
```
agneszitte, line 4: @eriklimakc you can use latest 5.4.8 stable version

agneszitte:
> @agneszitte @kazo0 Should we remove MacOS and Windows from the Platforms folder, launchSettings.json (only Windows), and from `<TargetFrameworks>`, since this app is Android and iOS only?
> 
> https://github.com/unoplatform/Uno.Samples/blob/19d039de6cdfa48d343ac0913f591d99481c5b50/UI/MediaGallery/src/MediaGallerySample/MediaGallery/MediaFileType.cs#L1

@eriklimakc I think we can keep only the platforms that are needed here. What do you think @kazo0 ?
Also as the class is supposed to work on Catalyst as well, shouldn't it be possible to keep MacOS ?
I did not check all the details yet, but would like your first thoughts @eriklimakc 

image

eriklimakc:
> @eriklimakc I think we can keep only the platforms that are needed here. What do you think @kazo0 ? Also as the class is supposed to work on Catalyst as well, shouldn't it be possible to keep MacOS ? I have not checked all the details yet, but I would like your first thoughts on it @eriklimakc please
> 

@agneszitte Oh yes, I didn't realise that `__IOS__` would also include mac. So yes, I think only Windows should be removed then. Right @kazo0?

agneszitte:
> > @eriklimakc I think we can keep only the platforms that are needed here. What do you think @kazo0 ? Also as the class is supposed to work on Catalyst as well, shouldn't it be possible to keep MacOS ? I have not checked all the details yet, but I would like your first thoughts on it @eriklimakc please
> 
> @agneszitte Oh yes, I didn't realise that `__IOS__` would also include mac. So yes, I think only Windows should be removed then. Right @kazo0?

@eriklimakc no `__IOS__` on this one normally does not include mac
cf. https://platform.uno/docs/articles/platform-specific-csharp.html#if-conditionals
image

eriklimakc:
> @eriklimakc no `__IOS__` on this one normally does not include macos/catalyst cf. https://platform.uno/docs/articles/platform-specific-csharp.html#if-conditionals

@agneszitte  That is funny, because when I change that dropdown to M⁠acCatalyst it continues showing the code as enabled:

image

image

While if I choose Windows the code shows as not enabled/ignored:

image


cc @kazo0

agneszitte:
This PR will need to be rebased with latest and updated to latest 5.5 stable version
Related new issue: https://github.com/unoplatform/Uno.Samples/issues/848

eriklimakc:
Can someone test iOS, please?

@agneszitte @kazo0

PR #875, opened 2025-03-31
docs-fix-links-and-lintings and part 1 of todo app tutorial creation

DevTKSS:
two nested Readme now Overview to be included in workshops in the future
Seperated just src links to be code-csharp blocks allowing dynamically showing src code in the docs along with its explanation.


DevTKSS:
Related to PR #763

I am currently checking, bugfixing and upgrading the ToDo App to mainly creating a more than source link tutorial for its contents.

Since for that its definitely best to check and fix src I am updating it, to current uno extensions standard and minimizing the warnings (better user experience etc.), so it would be good to know if that's maybe meant by this related PR PackageReference is meaning the named one or another thing instead needs to be applyed.

Would add the tutorial in a seperate PR then.

DevTKSS:
@kazo0 this here

reference/ToDo/doc/localize-by-model.md
```diff
@@ -0,0 +1,17 @@
...
+
+[!code-xaml[](../src/ToDo/Views/HomePage.xaml)]
+
+## Code-Behind of Home Page
+
+[!code-csharp[](../src/ToDo/Views/HomePage.xaml.cs)]
+
+## View Model of Home Page
+
+[!code-csharp[](../src/ToDo/Presentation/HomeViewModel.cs)]
```
kazo0, line 17: I don't think having docs pages here is particularly useful if we are just displaying application code and nothing else

kazo0: Same goes for most of the other new .md files in this directory

reference/ToDo/src/ToDo/Platforms/Desktop/Program.cs
```diff
@@ -7,8 +7,6 @@ public class Program
     [STAThread]
     public static void Main(string[] args)
     {
-        App.InitializeLogging();
-
```
kazo0, line 11: Why remove?

kazo0: Oh I see, Logging Extensions used

reference/TubePlayer/README.md
```diff
@@ -30,16 +30,16 @@ The TubePlayer app utilizes the latest Uno Platform features, including:
...
-- [Color overrides](https://platform.uno/docs/articles/external/uno.themes/doc/material-getting-started.html#manual-color-overrides)
-- [DSP import](https://platform.uno/docs/articles/external/uno.themes/doc/material-dsp.html)
-- [Material design theme](https://platform.uno/docs/articles/external/uno.themes/doc/material-getting-started.html)
-- [Toolkit](https://platform.uno/docs/articles/external/uno.toolkit.ui/doc/getting-started.html)
-- [Resizetizer](https://platform.uno/docs/articles/external/uno.resizetizer/doc/using-uno-resizetizer.html)
+- [Color overrides](https://aka.platform.uno/docs/articles/external/uno.themes/doc/material-getting-started.html#manual-color-overrides)
+- [DSP import](https://aka.platform.uno/docs/articles/external/uno.themes/doc/material-dsp.html)
+- [Material design theme](https://aka.platform.uno/docs/articles/external/uno.themes/doc/material-getting-started.html)
+- [Toolkit](https://aka.platform.uno/docs/articles/external/uno.toolkit.ui/doc/getting-started.html)
+- [Resizetizer](https://aka.platform.uno/docs/articles/external/uno.resizetizer/doc/using-uno-resizetizer.html)
```
kazo0: Have we validated that these are valid URLs? We don't have aka links for these

reference/ToDo/doc/localize-by-model.md
DevTKSS: @kazo0 you mentioned it to be a possible reference for custom Auth and as its using msal packages from what I saw in the Auth Service, but this sample only has [this entry in the Uno Docs](https://platform.uno/docs/articles/external/uno.samples/doc/samples.html#todo-app)
And [the ToDo App Blog post I could assume this was coming from](https://platform.uno/docs/articles/get-started-next-steps.html#building-a-todo-app-with-uno-platform-free) is still introducing us to the multi head setup 👀 
and its also not even what is shown here 🤷 
so the target would have been, to link the appropriate files and add some explaination to each, like what we are seeing and could do with this...

But if we dont want this, feel free to tell what of this you would like to keep, if some or I can also just close this PR and delete the branch.
Could you maybe then check on those other two PR's in this repo, what about those so the stale bot doesn't need to act on them? Thanks 👍 

- https://github.com/unoplatform/Uno.Samples/pull/891
- https://github.com/unoplatform/Uno.Samples/pull/892

reference/TubePlayer/README.md
DevTKSS: reverted those changes, thanks for the heads up 👍 I did forget to validate them

PR #907, opened 2026-02-03
chore: bump to 6.5.29

booskander:
**GitHub Issue:** closes #

## PR Type:
Bumping to new SDK version

## What is the current behavior? 🤔
A few warnings logged in runtime for Commerce app, but everything builds fine.
- In commerce some navigation issues but are known 

## PR Checklist ✅

Please check if your PR fulfills the following requirements:

- [ ] 📝 Commits must be following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification.
- [ ] 🧪 Added [Runtime tests, UI tests, or a manual test sample](https://github.com/unoplatform/uno/blob/master/doc/articles/uno-development/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] 📚 Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md) (for bug fixes / features)
- [ ] 🖼️ Validated PR `Screenshots Compare Test Run` results.
- [ ] ❗ Contains **NO** breaking changes

### DIFF
--- a/reference/Commerce/src/global.json
+++ b/reference/Commerce/src/global.json
@@ -1,7 +1,7 @@
 {
   // To update the version of Uno please update the version of the Uno.Sdk here. See https://aka.platform.uno/upgrade-uno-packages for more information.
   "msbuild-sdks": {
-    "Uno.Sdk": "6.3.28"
+    "Uno.Sdk": "6.5.29"
   },
   "sdk": {
     "allowPrerelease": false
--- a/reference/Counter/CSharp-MVUX/global.json
+++ b/reference/Counter/CSharp-MVUX/global.json
@@ -1,7 +1,7 @@
 {
   // To update the version of Uno please update the version of the Uno.Sdk here. See https://aka.platform.uno/upgrade-uno-packages for more information.
   "msbuild-sdks": {
-    "Uno.Sdk": "6.3.28"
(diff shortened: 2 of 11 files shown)


reference/Commerce/src/global.json
```diff
@@ -1,7 +1,7 @@
 {
   // To update the version of Uno please update the version of the Uno.Sdk here. See https://aka.platform.uno/upgrade-uno-packages for more information.
   "msbuild-sdks": {
-    "Uno.Sdk": "6.3.28"
+    "Uno.Sdk": "6.5.29"
```
nickrandolph, line 4: Can we move the global.json to the root folder so we don't need to update it is so many places?

agneszitte: @nickrandolph (cc @kazo0) One thing to keep in mind is that this would require people downloading a single reference sample to rely on an external root-level global.json to know which version to use.

It would also likely mean some CI changes, I imagine. I’m not fully convinced this adds much value compared to a quick find-and-replace in VS Code at the root level when it is time to bump to a new version.

rajamatt (approved):

PR #918, opened 2026-06-11
feat(uno-crm): modernize the CRM sample (responsive, mobile polish, shared nav, app icon & splash)

Jen-Uno:
Updates the **Uno CRM** sample (`studio/uno-crm`). Relates to unoplatform/studio.live#2416.

### 1. Responsive layout via the Toolkit
Replaced the `VisualStateManager` + `AdaptiveTrigger` desktop/mobile switch on all four pages with the Uno Toolkit `ResponsiveExtension` (`{utu:Responsive}`), bound to a single app-wide `DefaultResponsiveLayout` (crossover at **900px**, preserving the original `MinWindowWidth`). Adds the `Toolkit` UnoFeature.

### 2. Mobile layout polish
- Honor the device **safe area** (`utu:SafeArea`) so content clears the notch/status bar and the floating nav clears the home indicator.
- Removed the hardcoded **fake status bar** ("9:41" + fake signal/wifi/battery) from the Dashboard.
- Unified the four page headers to the same **title + subtitle + standard action button** pattern, and unified content padding/spacing.

### 3. Shared bottom navigation + pressed state
- Folded the bottom nav (previously duplicated verbatim across all four pages) into a reusable **`BottomNavBar`** control with an `ActiveTab` property.
- Fixed the **gray-rectangle pressed state** — the default WinUI button pressed background — replacing it with a rounded, theme-aware highlight. Same fix applied to the desktop sidebar nav style.

### 4. App icon + splash screens (Uno.Resizetizer)
- Configured the app **icon**, **regular splash**, and an **extended splash screen** (`utu:ExtendedSplashScreen` hosted in a new `Shell`) for all platforms via Resizetizer, on a `#151515` background.

## Verification
- Builds clean on **Desktop, iOS, and WASM** (0 errors).
- Verified on the **iOS simulator**: home-screen icon, native splash, extended splash, and the responsive mobile layout; verified the responsive desktop ↔ mobile switch and the nav pressed state in the browser at narrow/wide widths.

## Notes
- The app icon source is a 240×240 PNG, so it is upscaled for large targets (e.g. the 1024px iOS icon) and is slightly soft up close — a vector source would be sharper.
- Android wasn't built locally (needs JDK 17 in this environment), but the Resizetizer config and `MainActivity` init are platform-agnostic.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


studio/uno-crm/UnoCRM/UnoCRM.csproj
```diff
@@ -23,8 +23,27 @@
...
   </PropertyGroup>
+
+  <!--
+    App icon + splash screen via Uno.Resizetizer (all platforms).
+    Source is a bar-chart logo (transparent corners) on a #151515 dark background,
+    so the background colour fills the corners/adaptive-icon background seamlessly.
+  -->
+  <PropertyGroup>
+    <UnoIconForegroundFile>Assets/Icons/icon_foreground.svg</UnoIconForegroundFile>
+    <UnoIconColor>#151515</UnoIconColor>
```
carldebilly: **HIGH** — `UnoIconColor` is not a property the Uno SDK reads. In `unoplatform/uno`, `src/Uno.Sdk/targets/Uno.DefaultItems.Resizetizer.targets` maps `Color="$(UnoIconBackgroundColor)"` onto the `UnoIcon` item, defaulting to `#00000000` (transparent) when unset — `UnoIconColor` appears nowhere in the SDK targets or in uno.resizetizer's code, only in one erroneous example in its docs (the property list in that same doc says `UnoIconBackgroundColor`).

Impact is masked here because `icon.svg` is itself a solid `#151515` background, but the dead property means the Android adaptive-icon background layer (the one target not built locally) falls back to transparent.

Suggested fix: rename to `<UnoIconBackgroundColor>#151515</UnoIconBackgroundColor>` and verify the generated Android adaptive icon.

studio/uno-crm/UnoCRM/ContactsPage.xaml
```diff
@@ -173,24 +211,25 @@
...
+                      ItemsSource="{Binding Regions}"
                       SelectedIndex="0"
-                      SelectionChanged="RegionFilter_SelectionChanged">
-              <ComboBoxItem Content="All Regions" />
-              <ComboBoxItem Content="North America" />
-              <ComboBoxItem Content="Europe" />
-              <ComboBoxItem Content="APAC" />
-              <ComboBoxItem Content="LATAM" />
-            </ComboBox>
+                      SelectionChanged="RegionFilter_SelectionChanged" />
```
carldebilly, line 215: **HIGH** — `SelectedIndex="0"` is set at XAML parse time, but `ItemsSource="{Binding Regions}"` only resolves once `DataContext` is assigned after `InitializeComponent`. Under WinUI semantics the index is coerced to `-1` on the then-empty ComboBox and is not re-applied when the items arrive, so the combos can render blank on first load — and the behavior varies per Uno target (only iOS/WASM were verified visually). Same pattern on `DesktopSegmentFilter` below and the two mobile combos.

Suggested fix: set the initial selection in code-behind after assigning `DataContext` (e.g. `combo.SelectedItem = Regions[0]`), or verify on Desktop/Android that the parse-time index survives.

studio/uno-crm/UnoCRM/ContactsPage.xaml
```diff
@@ -337,33 +377,22 @@
...
+                        ItemsSource="{Binding Regions}"
                         SelectedIndex="0"
-                        SelectionChanged="RegionFilter_SelectionChanged">
-                <ComboBoxItem Content="All Regions" />
-                <ComboBoxItem Content="North America" />
-                <ComboBoxItem Content="Europe" />
-                <ComboBoxItem Content="APAC" />
-                <ComboBoxItem Content="LATAM" />
-              </ComboBox>
+                        SelectionChanged="RegionFilter_SelectionChanged" />
```
carldebilly, line 381: Same parse-time `SelectedIndex="0"` vs late-bound `ItemsSource` concern as the desktop filter combos above — see that comment.

studio/uno-crm/UnoCRM/Controls/BottomNavBar.xaml.cs
```diff
@@ -0,0 +1,194 @@
...
+            "Home" => typeof(MainPage),
+            "Pipeline" => typeof(PipelinePage),
+            "Leads" => typeof(LeadsPage),
+            "Contacts" => typeof(ContactsPage),
+            _ => null,
+        };
+
+        if (target is not null)
+        {
+            FindFrame()?.Navigate(target);
```
carldebilly: **HIGH** — two related issues in the tab navigation:

1. **Upward coupling**: the reusable control in `UnoCRM.Controls` hard-codes the four page types, so the "shared" control depends on the app's pages (DIP inverted) — any page rename/addition means editing the control, and it can't be reused elsewhere. Exposing the targets as tab metadata or raising a `TabSelected` event the host handles would invert that.
2. **Back-stack growth**: `Frame.Navigate` on every tab tap grows the back stack without bound — on Android, the Back button replays the entire tab history. Tab-style navigation usually clears/trims the back stack (`Frame.BackStack.Clear()` after navigating).

Related low: `ActiveTab` is a stringly-typed DP whose valid domain (`Home|Pipeline|Leads|Contacts`) lives in three places (doc comment, VSM state names, this switch) and a typo no-ops silently — shared constants or an enum would make it sturdier.

studio/uno-crm/UnoCRM/Controls/BottomNavBar.xaml.cs
```diff
@@ -0,0 +1,194 @@
...
+        {
+            if (VisualTreeHelper.GetChild(parent, i) is ScrollViewer scrollViewer)
+            {
+                _scrollViewer = scrollViewer;
+                _lastOffset = scrollViewer.VerticalOffset;
+                scrollViewer.ViewChanged += OnHostViewChanged;
+                return;
+            }
+        }
+    }
```
carldebilly, line 118: **MEDIUM** — the auto-hide relies on an implicit structural contract ("first `ScrollViewer` sibling of my parent") that lives only in this comment. A page that wraps its ScrollViewer in a Border/Grid silently loses auto-hide with no diagnostic. Consider an optional `ScrollHost` dependency property (ElementName-bindable) with this sibling scan as the fallback, and/or stating the contract in the class-level doc comment. (Subscription lifecycle itself checks out — Loaded/Unloaded are symmetric, no leak across navigations.)

studio/uno-crm/UnoCRM/Shell.xaml.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// </summary>
+    private sealed class StartupLoadable : ILoadable
+    {
+        public bool IsExecuting { get; private set; } = true;
+
+        public event EventHandler? IsExecutingChanged;
+
+        public void Begin(DispatcherQueue dispatcher, TimeSpan delay)
+        {
+            _ = Task.Delay(delay).ContinueWith(_ =>
```
carldebilly: **MEDIUM** — two issues with the splash dismissal:

1. **Silent hang path**: the fire-and-forget `Task.Delay().ContinueWith(... TryEnqueue ...)` never observes failure — if `TryEnqueue` returns `false` (dispatcher shutting down/not ready) or the continuation faults, `IsExecuting` stays `true` forever and the app sits on the splash with no signal. Check the `TryEnqueue` return value (or use a `DispatcherQueueTimer` / `async` method awaiting `Task.Delay`) and flip `IsExecuting` defensively.
2. **Artificial 2.5 s delay**: every launch on every platform pays 2.5 s purely to make the splash perceptible. Samples get copied verbatim, so this teaches fake loading rather than gating on real readiness — consider tying it to actual startup work (first frame navigated) or shortening it with a clearly-named constant (`SplashMinimumDuration`) and a demo-only comment.

studio/uno-crm/UnoCRM/ContactsPage.xaml.cs
```diff
@@ -167,15 +174,9 @@ private static void SyncComboSelection(ComboBox? first, ComboBox? second, string
...
+        if (!Equals(combo.SelectedItem as string, value))
         {
-            if (entry is ComboBoxItem item
-                && item.Content is string text
-                && string.Equals(text, value, StringComparison.OrdinalIgnoreCase))
-            {
-                combo.SelectedItem = item;
-                return;
-            }
+            combo.SelectedItem = value;
```
carldebilly, line 213: **MEDIUM** — (re: the desktop↔mobile sync helpers and `ApplyFiltersAndRefreshMap` just below) the Visibility-based responsive switch keeps both layouts alive, so the filter refresh rebuilds the contacts layer and `ForceUpdate()`s **two** live Mapsui `MapControl`s (each with its own OSM tile layer) on every keystroke — doubled memory/tile traffic, and a per-keystroke UI stall on single-threaded WASM. The replaced `MemoryLayer`s are also never disposed. Additionally, the desktop↔mobile mirroring (`SyncComboSelection` + search-box sync) triggers the counterpart's changed-event, so each user action runs this refresh twice — the re-entrancy only terminates thanks to the equality guards.

Suggested direction: update only the currently visible map (and dispose replaced layers), debounce the search (~250 ms), and consider binding both layouts TwoWay to single `SelectedRegion`/`SearchQuery` properties so the sync helpers disappear.

studio/uno-crm/UnoCRM/ContactsPage.xaml.cs
```diff
@@ -18,12 +18,26 @@ public sealed partial class ContactsPage : Page, INotifyPropertyChanged
...
+            .Concat(_allContacts.Select(c => c.Segment).Distinct(StringComparer.OrdinalIgnoreCase))
+            .ToArray();
+
         DataContext = this;
         Loaded += ContactsPage_Loaded;
-
-        BuildSampleContacts();
     }
 
+    public string[] Regions { get; private set; } = [];
```
carldebilly: **LOW** — `Regions`/`Segments` are exposed as `string[]` with `private set` but only assigned in the constructor; get-only `IReadOnlyList<string>` would state intent (arrays expose mutation). Related: the `"All Regions"`/`"All Segments"` literals are duplicated between the field initializers, this constructor, and the filter predicates — a rename desyncs the filtering; consider `private const string` for both.

studio/uno-crm/UnoCRM/Controls/PipelineCard.xaml.cs
```diff
@@ -0,0 +1,52 @@
...
+    public string Meta { get => (string)GetValue(MetaProperty); set => SetValue(MetaProperty, value); }
+    public Brush? Accent { get => (Brush?)GetValue(AccentProperty); set => SetValue(AccentProperty, value); }
+    public Brush? MetaBrush { get => (Brush?)GetValue(MetaBrushProperty); set => SetValue(MetaBrushProperty, value); }
+    public bool IsWon { get => (bool)GetValue(IsWonProperty); set => SetValue(IsWonProperty, value); }
+
+    public Visibility DotVisibility(bool isWon) => isWon ? Visibility.Collapsed : Visibility.Visible;
+
+    public Visibility CheckVisibility(bool isWon) => isWon ? Visibility.Visible : Visibility.Collapsed;
+
+    public Windows.UI.Text.FontWeight MetaWeight(bool isWon) => isWon ? FontWeights.SemiBold : FontWeights.Normal;
```
carldebilly: **LOW** — `DotVisibility`/`CheckVisibility`/`MetaWeight` are `public` but only consumed by the control's own `x:Bind` (whose generated code lives in the same partial class), so `private` suffices. The two visibility helpers also re-implement the bool→`Visibility` conversion that `x:Bind` performs implicitly.

.claude/lessons.md
```diff
@@ -0,0 +1,404 @@
...
+  `dotnet run --no-build -f net10.0-browserwasm`, poll `http://localhost:5000`, then
+  drive with Playwright. The app renders to a `<canvas>`, so navigate with
+  **coordinate clicks**, and capture at specific viewport widths (and
+  `colorScheme: 'dark'`) to validate responsive breakpoints.
+- **Android** builds need JDK 17; Skia desktop is windowless from a headless shell.
+
+---
+
+*Generated from the change history on branch `dev/jenny/update-uno-crm-sample`
+(PR for unoplatform/studio.live#2416).*
```
carldebilly: **LOW** — this references the private `unoplatform/studio.live` repo in a public OSS repo: a dead link for external readers and a minor internal-linkage disclosure. Suggest referencing the public PR (unoplatform/Uno.Samples#918) instead.

carldebilly (approved):
Really nice modernization overall — the ResponsiveExtension migration, the BottomNavBar/PipelineCard extractions, the Resizetizer SVG icon/splash setup, and the lessons.md write-up are all solid, and the XAML↔code-behind wiring (handlers, x:Name targets, ThemeResource keys, Toolkit feature flag) checks out. Approving with a handful of findings worth a follow-up — 3 high, 4 medium, a few low.

A few notes on code that sits outside the diff hunks (so no inline anchor):

- **ApplicationId** (`UnoCRM.csproj`): still `com.companyname.unosfa` while the app is branded Uno CRM. Package identity is painful to change after any distribution — worth renaming (e.g. `com.unoplatform.unocrm`) while the sample is young.
- **LeadsPage chart colors**: the LiveCharts palettes are resolved once in the constructor via `Application.Current.Resources.TryGetValue` (with hardcoded fallbacks like `SKColors.White`). A runtime light/dark toggle — which this sample just gained — leaves the charts painted with the stale theme. Consider rebuilding on `ActualThemeChanged`.
- **OSM tile layer** (`ContactsPage`): `OpenStreetMap.CreateTileLayer()` is called without an explicit user-agent; OSM throttles/blocks default UAs, and offline the map goes silently blank. Passing a UA string and a one-line comment about the degraded mode would do for a sample.
- **Desktop sidebar duplication**: the sidebar (logo + nav + styles) is still quadruplicated across the four pages while the bottom nav got extracted — the PR's own lessons.md (#8/#11) prescribes exactly that extraction. A `SidebarNav` control mirroring `BottomNavBar` would be a natural follow-up.
- **lessons.md location**: great content; note that `.claude/` at repo root reads as agent config rather than contributor docs, so people browsing `studio/uno-crm` won't find it.

Verified clean: no secrets, SVG assets are pure vector (no scripts/external refs), no new warning suppressions, no blocking synchronization primitives (WASM-safe), and BottomNavBar's Loaded/Unloaded subscription lifecycle is symmetric (no leak across navigations).

studio/uno-crm/UnoCRM/UnoCRM.csproj
Jen-Uno: Fixed in 02f900f2 — renamed to `UnoIconBackgroundColor`. Confirmed via the generated `UnoImage.inputs`: the app-icon record now resolves `Color=#151515` (was `#00000000`), so the Android adaptive-icon background layer is no longer transparent. Thanks for the precise SDK pointer.

studio/uno-crm/UnoCRM/ContactsPage.xaml
Jen-Uno: Fixed in 02f900f2 — the four combos now receive their initial selection in code-behind (`ContactsPage_Loaded`, after `DataContext`/`ItemsSource` resolve), so an index coerced to `-1` on the then-empty source no longer leaves them blank. An `_initialized` guard suppresses the load-time `SelectionChanged` churn. The XAML `SelectedIndex="0"` is left as a harmless fast-path on targets where it survives.

studio/uno-crm/UnoCRM/ContactsPage.xaml
Jen-Uno: Covered by the same code-behind initial-selection fix in 02f900f2 — it applies to all four combos (desktop + mobile).

studio/uno-crm/UnoCRM/Controls/BottomNavBar.xaml.cs
Jen-Uno: **Back-stack growth** fixed in 02f900f2 — tab navigation now clears `Frame.BackStack` after a successful `Navigate`, so Android Back no longer replays tab history.

**Upward coupling**: kept deliberately for this single-app sample — the control lives with the app (`UnoCRM.Controls`) and the self-contained page→type switch reads more clearly here than routing navigation through each page via an event. Agreed that a `TabSelected` event / tab metadata is the right inversion for a genuinely reusable library control.

**Stringly-typed `ActiveTab`**: noted — left as-is since the domain is small and the VSM state names (XAML) can't share a C# enum without a converter; the trade-off didn't feel worth the indirection for the sample.

studio/uno-crm/UnoCRM/Controls/BottomNavBar.xaml.cs
Jen-Uno: Addressed in 02f900f2 — added an optional `ScrollHost` dependency property (ElementName-bindable); the "first `ScrollViewer` sibling of my parent" scan is now the fallback. Stated the auto-hide host contract in the class-level doc comment so a page that nests its scroller knows to set `ScrollHost`.

studio/uno-crm/UnoCRM/Shell.xaml.cs
Jen-Uno: Fixed in 02f900f2:
1. **Silent hang** — the loader now `await`s `Task.Delay` then completes via `TryEnqueue`, falling back to flipping `IsExecuting` inline if the enqueue fails, so it can never stay stuck on the splash.
2. **Artificial delay** — the 2.5 s is now a named `SplashMinimumDuration` constant with a demo-only comment noting a production app should gate the reveal on real readiness (first navigation / data) rather than a fixed delay.

studio/uno-crm/UnoCRM/ContactsPage.xaml.cs
Jen-Uno: Partially addressed in 02f900f2 — added `_initialized` / `_syncing` guards so the desktop↔mobile mirroring no longer raises a second `SelectionChanged`, i.e. a change now refreshes **once** (the re-entrancy you flagged), and the load-time selection no longer storms `ApplyFilters`.

Left within the sample's scope (≈85 in-memory points): both maps stay live so the responsive switch never shows a stale/blank map, and the search isn't debounced. Noted the dispose-replaced-`MemoryLayer`s, ~250 ms search debounce, and single-source-of-truth `TwoWay` binding as the right production direction.

studio/uno-crm/UnoCRM/ContactsPage.xaml.cs
Jen-Uno: Done in 02f900f2 — `Regions`/`Segments` are now get-only `IReadOnlyList<string>`, and the `"All Regions"` / `"All Segments"` literals are `private const` used by the field initializers, the derivation, and the filter predicates.

studio/uno-crm/UnoCRM/Controls/PipelineCard.xaml.cs
Jen-Uno: Made the three helpers `private` in 02f900f2 (x:Bind's generated code is in the same partial class). Kept the two visibility helpers: the check binds `IsWon` directly, but the dot needs `!IsWon`, which x:Bind can't negate inline without a function/converter — so the explicit helper is the clearest expression of that.

.claude/lessons.md
Jen-Uno: Fixed in 02f900f2 — now references the public `unoplatform/Uno.Samples#918` instead of the private repo.

Jen-Uno:
Thanks for the thorough review @carldebilly 🙏 — all 10 comments addressed in 02f900f2 (replied inline + resolved):

**Fixed**
- **HIGH** `UnoIconColor` → `UnoIconBackgroundColor` (verified the generated icon now resolves `Color=#151515`, not transparent).
- **HIGH** ComboBox `SelectedIndex` vs late-bound `ItemsSource` — initial selection now set in code-behind after binding (all four combos).
- **HIGH** BottomNavBar back-stack growth — clears `Frame.BackStack` after tab navigation.
- **MEDIUM** Added an optional `ScrollHost` DP (sibling scan as fallback) + documented the auto-hide contract.
- **MEDIUM** Splash now dismisses defensively (never stuck) with a named demo-only duration constant.
- **MEDIUM** Filter combo re-entrancy — `_initialized`/`_syncing` guards so a change refreshes once.
- **LOW** `IReadOnlyList` + `const` for the "All …" literals; PipelineCard helpers `private`; lessons.md now links the public PR.

**Kept by design (sample scope), with reasoning in-thread**
- BottomNavBar's page→type coupling (app-specific control) and stringly-typed `ActiveTab`.
- Both maps kept live for the responsive switch + no search debounce (≈85 in-memory points) — noted as the right production direction.

Note: changes are build-verified on WASM; I couldn't visually verify rendered output this round, so a quick look on Desktop/Android for the combo selection and adaptive icon would be appreciated.

